PR #65, opened 2014-07-31
Adding multi-line example

kentcdodds:
Because some people don't know that `cson` allows multi-line strings.


raxell:
Does indentation in multiline snippets need to use the `\t` char?
Which of these is better?
```coffee
  'body': """
    if ($1) {
      $2
    }
  """
```
or
```coffee
  'body': """
    if ($1) {
    \t$2
    }
  """
```

PR #82, opened 2014-11-20
How to find out the correct scope (selector)

pbihler:
Added findings from https://discuss.atom.io/t/how-do-you-configure-use-snippets/335/28


izuzak:
I cleaned this up a bit and merged -- sorry for the delay and thanks for opening this. :bow:

Issue #59, opened 2014-07-02
snippets:expand on tab (default) inside PHP variable triggered


IwanLuijksQuestMedia:
When I define a variable in PHP, and my cursor is within the variable '$address' (between the double 's'), or right after the variable '$address' and I hit [tab] the 'snippets:expand' command is triggered:
![schermafbeelding 2014-07-02 om 13 56 49](https://cloud.githubusercontent.com/assets/5575320/3455430/1bccd038-01e1-11e4-8a4f-df06a5647b9a.png)
![schermafbeelding 2014-07-02 om 13 57 02](https://cloud.githubusercontent.com/assets/5575320/3455451/5c1744d4-01e1-11e4-9a78-025f57eb70bf.png)

This is unexpected behavior, since I just wanted a tab being inserted.

From within the '$address'-variable it isn't very likely to occur (the double 's' expansion to ```<s></s>```), but hitting [tab] right after the variable definition is actually something I do a lot, and probably most do.

Maybe the **default** keybinding of [tab] to 'snippets:expand' could be changed to something else?

Mac OS X 10.8.5
Atom Version 0.106.0 (0.106.0)
Snippets package: 0.46.0

maxbrunsfeld:
Ok, I believe this is fixed as of version 0.70.0. Please let us know if there are still bugs. Thanks!

PR #99, opened 2015-01-15
Don't trigger expansion when prefix is part of a larger word

maxbrunsfeld:
Currently, if a snippet prefix precedes the cursor, snippet expansion is triggered by the `tab` key. This PR *prevents* triggering snippet expansion if the snippet prefix preceding the cursor is part of a larger word, as defined by the `editor.nonWordCharacters` config setting. 

I'm not sure if this is the right thing to do, but I think it addresses issues raised by @dsandstrom in #59.

### DIFF
--- a/lib/snippets.coffee
+++ b/lib/snippets.coffee
@@ -186,7 +186,9 @@ module.exports =
     cursors = editor.getCursors()
     for cursor in cursors
       prefixStart = cursor.getBeginningOfCurrentWordBufferPosition({wordRegex})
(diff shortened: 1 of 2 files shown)


kevinsawicki:
Yeah, this sounds good to me, no common cases just to mind that this would "break" and I imagine this makes things more consistent with other editors people may be using snippets in.

maxbrunsfeld:
Ok, I think I'm going to say that this Fixes #59.

maxbrunsfeld:
Oops after doing some testing, this doesn't work in some cases. I'm going to fix it up.

PR #105, opened 2015-02-01
Added further explanation for the outermost key

hitautodestruct:
Added further explanation that details you need to specify the file type for the snippet to appear in. If there is more to this then it should be documented.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Snippet files can be either `.json` or `.cson`.
     'body': 'console.log(${1:"crash"});$2'
 ```
 
-The outermost keys are the selectors where this snippets should be active.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,7 +18,7 @@ Snippet files can be either `.json` or `.cson`.
     'body': 'console.log(${1:"crash"});$2'
 ```
 
-The outermost keys are the selectors where this snippets should be active.
+The outermost keys are the selectors where this snippets should be active. If you would want the snippet to be active in python, for example, you would use `'.source.py'`.
```
lee-dohm: This should be `.source.python`. c.f.:

https://github.com/atom/language-python/blob/master/snippets/language-python.cson#L1

hitautodestruct:
@lee-dohm Woops, all fixed now :)

kevinsawicki:
:ship:

PR #106, opened 2015-02-05
[WIP] Add Service API To Snippets

joefitzgerald:
:rotating_light: WIP :rotating_light:

Not ready for merge yet:

- [x] Add API
- [ ] Add tests
- [ ] Update autocomplete-plus to consume the API and validate it has the right surface area

:rotating_light: WIP :rotating_light:

### DIFF
--- a/lib/snippets.coffee
+++ b/lib/snippets.coffee
@@ -273,9 +273,18 @@ module.exports =
   addExpansion: (editor, snippetExpansion) ->
     @getExpansions(editor).push(snippetExpansion)
 
-  insert: (snippet, editor=atom.workspace.getActiveTextEditor(), cursor=editor.getLastCursor()) ->
+  insert: (snippet, editor = atom.workspace.getActiveTextEditor(), cursor = editor.getLastCursor()) ->
     if typeof snippet is 'string'
       bodyTree = @getBodyParser().parse(snippet)
       snippet = new Snippet({name: '__anonymous', prefix: '', bodyTree, bodyText: snippet})
 
     new SnippetExpansion(snippet, editor, cursor, this)
+
+  provideSnippets: ->
+    isLoaded = -> @loaded
+    snippets = this
+    return s =
+      loaded: isLoaded.bind(snippets)
+      insert: @insert.bind(snippets)
+      getSnippets: @getSnippets.bind(snippets)
+      getPrefixText: @getPrefixText.bind(snippets)
--- a/package.json
+++ b/package.json
@@ -18,5 +18,13 @@
     "season": "^5.0.4",
     "temp": "~0.8.0",
     "underscore-plus": "^1.0.0"
+  },
+  "providedServices": {
+    "snippets": {
+      "description": "Snippets are text shortcuts that can be expanded to their definition.",
+      "versions": {
+        "0.73.0": "provideSnippets"
+      }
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -18,5 +18,13 @@
     "season": "^5.0.4",
     "temp": "~0.8.0",
     "underscore-plus": "^1.0.0"
+  },
+  "providedServices": {
+    "snippets": {
+      "description": "Snippets are text shortcuts that can be expanded to their definition.",
+      "versions": {
+        "0.73.0": "provideSnippets"
```
benogle: This should probably be 1.0.0

lib/snippets.coffee
```diff
@@ -273,9 +273,18 @@ module.exports =
...
     new SnippetExpansion(snippet, editor, cursor, this)
+
+  provideSnippets: ->
+    isLoaded = -> @loaded
+    snippets = this
+    return s =
+      loaded: isLoaded.bind(snippets)
+      insert: @insert.bind(snippets)
+      getSnippets: @getSnippets.bind(snippets)
+      getPrefixText: @getPrefixText.bind(snippets)
```
benogle: I think we should leave this super minimal and just provide the `insert` method. It's really all autocomplete needs to do its thing.

package.json
joefitzgerald: Was just following the pattern @maxbrunsfeld seemed to establish in status-bar.

lib/snippets.coffee
joefitzgerald: These additional things cater to `autocomplete-snippets` needs. And isLoaded is required to allow testing to be possible.

package.json
benogle: status-bar wasnt the ideal scenario. We want to version the service independently from the package version.

joefitzgerald: :thumbsup:

benogle: After talking to @maxbrunsfeld we decided this should probably start at `0.1.0` rather than `1.0.0`.

joefitzgerald: Sounds good.

lib/snippets.coffee
benogle: autocomplete-snippets only uses `loaded` in specs and it get the snippets via `atom.config`. Getting the snippets from config provides less surface area and is much more flexible than the `getSnippets` method. Looks like `getPrefixText` is not used anywhere.

Maybe loaded stays, but with a new name for more clarity `allSnippetsLoaded` or `bundledSnippetsLoaded`.

benogle:
I can take this over as I am tinkering with snippets in autocomplete plus

lib/snippets.coffee
joefitzgerald: getPrefixText would be used to replace the default prefix that is provided by autocomplete-plus in the suggestion request: https://github.com/atom-community/autocomplete-snippets/blob/7e183365030cfa96586077d0e7eb339d5b52eace/lib/snippets-provider.coffee#L9. I would assume the snippets package has better logic for determining the prefix of a snippet than we do, but perhaps this is moot given the simplicity of most prefixes (they are generally `[\w]*`).

benogle: I would rather the prefix were generated from the autocomplete package or by asking the providers.

PR #108, opened 2015-02-16
Support escaped $ symbol in snippet body

sergeche:
Allow escaped `$` in snippet body to allow strings that look like a tabstop.
E.g. output `\$1` as `$1`, not as `\` + tabstop

PS: I was unable to run Jasmine specs because of `Module did not self-register` error, but I tested it with online PEG playground


kevinsawicki:
Looks like the spec you added is failing on Travis: https://travis-ci.org/atom/snippets/builds/51000360

kevinsawicki:
>  I was unable to run Jasmine specs because of Module did not self-register error

Run `apm rebuild` from the root of the repository, and it should resolve this.

sergeche:
Oh, sorry, there was a typo in specs. Should work fine now

kevinsawicki:
Thanks for this :ship:

PR #110, opened 2015-02-20
Add service API 

benogle:
This finishes what @joefitzgerald started in #106. It just adds specs and reduces the api footprint a bit. 

Closes #106


joefitzgerald:
LGTM.

PR #113, opened 2015-03-04
Terminate snippets when cursor gets destroyed

as-cii:
Cursors were being already tracked, but only when moving them. When `Esc` gets pressed, cursors are not moved: editor, in facts, consolidates selections, thus destroying some of the cursors. This solves https://github.com/atom/atom/issues/4821

/cc: @izuzak @nathansobo


nathansobo:
:beetle: :gun: Good thinking.

PR #120, opened 2015-03-19
Fix same index tab stops exiting prematurely

### LINKED ISSUE #118: Snippets with same index tab stops exit prematurely
This snippet has the same index for its first and last tab stops.

```cson
'.text.html.basic':
  'Spacebars Block':
    'prefix': 'sbblock'
    'body': '''
      {{#${1:block} ${2:context}}}
        ${3:body}
      {{/${1:block}}}'''
```

Tabbing to the the second tab stop works, but tabbing to the third does not. It just inserts a tab as if all tab stops had been reached.

I tracked the cause down to a [recent commit][1517063c]. It seems that the `newSelection` variable is only used when a snippet has more than one tab stop with the same index. However, when this `newSelection` is destroyed (i.e. on the next <kbd>Tab</kbd> press) it destroys the `SnippetExpansion` associated with it. This causes the following <kbd>Tab</kbd> press to [abort its key binding][SL39] because there is no `SnippetExpansion` anymore.

Commenting out [that line][SEL74] fixes this issue, and in my testing it doesn't resurrect the issue that the commit fixed ([atom#4821][ATOM4821]) after following the [reproductions steps][REPO4821].

The funny thing is, you can actually follow those same reproduction steps to reproduce this new issue.

[1517063c]: https://github.com/atom/snippets/commit/1517063c8cd2c8e2be70cdd8dc8a2f94ddf07b2b
[SL39]: https://github.com/atom/snippets/blob/1517063c8cd2c8e2be70cdd8dc8a2f94ddf07b2b/lib/snippets.coffee#L39
[SEL74]: https://github.com/atom/snippets/blob/1517063c8cd2c8e2be70cdd8dc8a2f94ddf07b2b/lib/snippet-expansion.coffee#L74
[ATOM4821]: https://github.com/atom/atom/issues/4821
[REPO4821]: https://github.com/atom/atom/issues/4821#issuecomment-71965142

jordanbtucker:
After testing, I'm convinced that destruction of `newSelection` does not need to be handled. Handling cursor change should be enough, and it doesn't have the side effect of breaking the tab stops.

Fixes #118

### DIFF
--- a/lib/snippet-expansion.coffee
+++ b/lib/snippet-expansion.coffee
@@ -17,7 +17,7 @@ class SnippetExpansion
         @cursor.selection.insertText(snippet.body, autoIndent: false)
       if snippet.tabStops.length > 0
         @subscriptions.add @cursor.onDidChangePosition (event) => @cursorMoved(event)
-        @subscriptions.add @cursor.onDidDestroy => @destroy()
+        @subscriptions.add @cursor.onDidDestroy => @cursorDestroyed()
         @placeTabStopMarkers(startPosition, snippet.tabStops)
         @snippets.addExpansion(@editor, this)
         @editor.normalizeTabsInBufferRange(newRange)
@@ -28,6 +28,8 @@ class SnippetExpansion
     oldTabStops = @tabStopsForBufferPosition(oldBufferPosition)
     newTabStops = @tabStopsForBufferPosition(newBufferPosition)
     @destroy() unless _.intersection(oldTabStops, newTabStops).length
+
+  cursorDestroyed: -> @destroy() unless @settingTabStop
 
   placeTabStopMarkers: (startPosition, tabStopRanges) ->
     for ranges in tabStopRanges
@@ -71,7 +73,7 @@ class SnippetExpansion
         else
           newSelection = @editor.addSelectionForBufferRange(range)
           @subscriptions.add newSelection.cursor.onDidChangePosition (event) => @cursorMoved(event)
-          @subscriptions.add newSelection.cursor.onDidDestroy => @destroy()
+          @subscriptions.add newSelection.cursor.onDidDestroy => @cursorDestroyed()
           @selections.push newSelection
       markerSelected = true
 
--- a/spec/snippets-spec.coffee
+++ b/spec/snippets-spec.coffee
@@ -114,6 +114,7 @@ describe "Snippets extension", ->
               with placeholder ${1:test}
               without placeholder $1
               second tabstop $2
+              third tabstop $3
             """
 
           "large indices":
@@ -453,6 +454,20 @@ describe "Snippets extension", ->
         # this should insert whitespace instead of going through tabstops of the previous destroyed snippet
         simulateTabKeyEvent()
         expect(editor.lineTextForBufferRow(2).indexOf("  second")).toBe 0
+
+      it "moves to the second tabstop after a multi-caret tabstop", ->
+        editor.setCursorScreenPosition([0, 0])
+        editor.insertText('t9b')
+        simulateTabKeyEvent()
+        editor.insertText('line 1')
+
+        simulateTabKeyEvent()
+        editor.insertText('line 2')
+
+        simulateTabKeyEvent()
+        editor.insertText('line 3')
+
+        expect(editor.lineTextForBufferRow(2).indexOf("line 2 ")).toBe -1
 
     describe "when the snippet contains tab stops with an index >= 10", ->
       it "parses and orders the indices correctly", ->


50Wliu:
It's odd how this change didn't produce a test failure.  Could you by any chance add a test that would fail without this change?

jordanbtucker:
Test added.

as-cii:
My bad! Apparently it was me to introduce that issue: I hope this hasn't caused too many headaches to you. Sorry for that :pray: 

@jordanbtucker: Nice job, by the way! I think we can merge this very soon.

Do you think we can reuse existing fixtures in this new test you have written?

> After testing, I'm convinced that destruction of newSelection does not need to be handled. 

I am quite sure you're right here, but could you elaborate a bit this thought, please? Why don't we need to handle `newSelection` destruction?

Thank you very much @jordanbtucker! :rainbow:

jordanbtucker:
@as-cii No headaches here.

When you `:consolidate-selections`, Atom destroys all but the *last* cursor added. So handling the destruction of `@cursor` is sufficient for that case. I'm not sure else what would destroy `newSelection`'s cursor, but I added code to handle that and destroy the `SnippetExpansion` only if not `@settingTabStop`. Can you think of any issues with that?

I also refactored the code to use existing test fixtures.

lib/snippet-expansion.coffee
```diff
@@ -71,7 +71,8 @@ class SnippetExpansion
         else
           newSelection = @editor.addSelectionForBufferRange(range)
           @subscriptions.add newSelection.cursor.onDidChangePosition (event) => @cursorMoved(event)
-          @subscriptions.add newSelection.cursor.onDidDestroy => @destroy()
+          @subscriptions.add newSelection.cursor.onDidDestroy =>
+            @destroy() if not @settingTabStop
```
jordanbtucker: On new line so it's not read as:

```coffee
if not @settingTabStop
  @subscriptions.add newSelection.cursor.onDidDestroy => @destroy()
```

as-cii: Really minor, but you could use `unless` here.

as-cii:
> I also refactored the code to use existing test fixtures.

Awesome!

> When you :consolidate-selections, Atom destroys all but the last cursor added. So handling the destruction of @cursor is sufficient for that case. I'm not sure else what would destroy newSelection's cursor, but I added code to handle that and destroy the SnippetExpansion only if not @settingTabStop. Can you think of any issues with that?

Actually I think is what we should do on every cursor destroy event. This would also be consistent with how `cursorMoved` works. What do you think about something like:

```diff
diff --git a/lib/snippet-expansion.coffee b/lib/snippet-expansion.coffee
index 8207cc4..e28ad8c 100644
--- a/lib/snippet-expansion.coffee
+++ b/lib/snippet-expansion.coffee
@@ -17,7 +17,7 @@ class SnippetExpansion
         @cursor.selection.insertText(snippet.body, autoIndent: false)
       if snippet.tabStops.length > 0
         @subscriptions.add @cursor.onDidChangePosition (event) => @cursorMoved(event)
-        @subscriptions.add @cursor.onDidDestroy => @destroy()
+        @subscriptions.add @cursor.onDidDestroy => @cursorDestroyed()
         @placeTabStopMarkers(startPosition, snippet.tabStops)
         @snippets.addExpansion(@editor, this)
         @editor.normalizeTabsInBufferRange(newRange)
@@ -29,6 +29,8 @@ class SnippetExpansion
     newTabStops = @tabStopsForBufferPosition(newBufferPosition)
     @destroy() unless _.intersection(oldTabStops, newTabStops).length
 
+  cursorDestroyed: -> @destroy() unless @settingTabStop
+
   placeTabStopMarkers: (startPosition, tabStopRanges) ->
     for ranges in tabStopRanges
       @tabStopMarkers.push ranges.map ({start, end}) =>
@@ -71,8 +73,8 @@ class SnippetExpansion
         else
           newSelection = @editor.addSelectionForBufferRange(range)
           @subscriptions.add newSelection.cursor.onDidChangePosition (event) => @cursorMoved(event)
-          @subscriptions.add newSelection.cursor.onDidDestroy =>
-            @destroy() if not @settingTabStop
+          @subscriptions.add newSelection.cursor.onDidDestroy => @cursorDestroyed()
           @selections.push newSelection
       markerSelected = true
 
```

I think we should be good to go after completing this. Thank you very much @jordanbtucker :fire:

50Wliu:
After looking into this a bit, looks like this will also resolve atom/language-c#65 and https://discuss.atom.io/t/special-characters-in-snippets/15445 (which is now realize doesn't have to do with special characters).

jordanbtucker:
Those changes all look good. I haven't had a chance to get to this today, but I'll make the changes in an hour or so. Thanks for all the input.

as-cii:
@jordanbtucker: thank you very much for this! Awesome job here :100: 

@50Wliu: can you confirm this fixes the issue you've mentioned? If yes, could you close them please?

Thanks everyone! :+1:

50Wliu:
I'll check once 0.188.0 comes out-haven't gotten around to trying to build from source yet :wink:.

izuzak:
I just checked #118, atom/language-c#65 and https://discuss.atom.io/t/special-characters-in-snippets/15445 and all seem to be fixed.

as-cii:
> I just checked #118, atom/language-c#65 and https://discuss.atom.io/t/special-characters-in-snippets/15445 and all seem to be fixed.

That's great to know! Thanks @izuzak and @50Wliu :bow:

izuzak:
Just published snippets v0.85.0 and bumped the version in atom/atom. Thanks for the great work here @jordanbtucker and @as-cii. :surfer: :surfer:

PR #122, opened 2015-03-23
Add snippet to create snippets under a source definition

taylon:
Hi,

When I'm writing snippets it's really common to create many snippets under one source definition, so I added a snippet for that.

Not sure if snipk/"Atom Snippet Keys" is the best naming for that but it fits.

I also changed the placeholder for prefix because I think it would be easier to understand. Actually there is any reason why the key is not called "trigger" instead of "prefix"? I think it would be much easier to understand, if you guys think it is a good idea I can try to implement it maintaining compatibility, although I think that changing it at this point will probably cause more confusion than benefits.

### DIFF
--- a/lib/snippets.cson
+++ b/lib/snippets.cson
@@ -5,11 +5,20 @@
       {
         "${1:.source.js}": {
           "${2:Snippet Name}": {
-            "prefix": "${3:hello}",
+            "prefix": "${3:Snippet Trigger}",
             "body": "${4:Hello World!}"
           }
         }
       }$5
+    """
+
+  'Atom Snippet Keys':
+    prefix: 'snipk'
+    body: """
+      "${1:Snippet Name}": {
+        "prefix": "${2:Snippet Trigger}",
+        "body": "${3:Hello World!}"
+      }$4
     """
 
   'Atom Keymap':
@@ -28,8 +37,16 @@
     body: """
       '${1:.source.js}':
         '${2:Snippet Name}':
-          'prefix': '${3:hello}'
+          'prefix': '${3:Snippet Trigger}'
           'body': '${4:Hello World!}'$5
+    """
+
+  'Atom Snippet Keys':
+    prefix: 'snipk'
+    body: """
+      '${1:Snippet Name}':
+        'prefix': '${2:Snippet Trigger}'
+        'body': '${3:Hello World!}'$5
     """
 
   'Atom Keymap':


kevinsawicki:
> Actually there is any reason why the key is not called "trigger" instead of "prefix"?

I think because trigger better refers to the keybinding that causes the snippet to expand which defaults to `tab`, where prefix is used a bunch of places to refer to the text behind the cursor.

lib/snippets.cson
```diff
@@ -5,13 +5,22 @@
...
           "${2:Snippet Name}": {
-            "prefix": "${3:hello}",
+            "prefix": "${3:Snippet Trigger}",
             "body": "${4:Hello World!}"
           }
         }
       }$5
     """
 
+  'Atom Snippet Keys':
```
kevinsawicki: Maybe `Atom Snippet With No Selector`? This is a tricky name

taylon: Indeed it is...
 
Atom Snippet With No Selector sounds good, but what would be the prefix for that?

Actually, I was thinking about it, what if we do this:

```coffee
'.source.coffee':
  'Atom Snippet':
    prefix: 'snip'
    body: """
      ${1:'.${2:selector}':
        }'${3:Snippet Name}':
          'prefix': '${4:Snippet Trigger}'
          'body': '${5:Hello World!}'$5
    """
```

Instead of creating a new snippet we can make the fist line optional, so type 'snip' hit tab and it will highlight the whole line including the tabs in the next line, if you don't want the selector hit backspace and tab, if you want the selector just hit tab.

What do you think?

izuzak: Sorry for the delay in getting back to you on this, @taylon. You asked:

> Atom Snippet With No Selector sounds good, but what would be the prefix for that?

What do you think of `snipns`?

> Instead of creating a new snippet we can make the fist line optional, so type 'snip' hit tab and it will highlight the whole line including the tabs in the next line, if you don't want the selector hit backspace and tab, if you want the selector just hit tab.

That's an interesting idea. I think I'd find two snippets more intuitive than one in this case, though.

lib/snippets.cson
```diff
@@ -28,10 +37,18 @@
...
+          'prefix': '${3:Snippet Trigger}'
           'body': '${4:Hello World!}'$5
     """
 
+  'Atom Snippet Keys':
+    prefix: 'snipk'
+    body: """
+      '${1:Snippet Name}':
+        'prefix': '${2:Snippet Trigger}'
+        'body': '${3:Hello World!}'$5
```
izuzak: Should `'body': '${3:Hello World!}'$5` be `'body': '${3:Hello World!}'$4` ? (Notice the `$5` vs `$4`.)

lib/snippets.cson
taylon: Good one, I think snipns make sense.

About the second one, It's really hard to tell because I think it is very personal. I think that having one "customizable" snippet is better because once you have many snippets it's hard to remember them, so I rather have fewer snippets that can handle different situations than having one for each thing I wanna do. But I do agree that having two of them is more intuitive, specially in this case.

lib/snippets.cson
taylon: Yeah, it should be $4, it was a mistake.

izuzak:
Thanks for this, @taylon -- I made some minor changes in 17e4de9 based on our discussion. :bow:

PR #123, opened 2015-03-29
Handle escaped characters in tab-stop placeholder content

### LINKED ISSUE #60: Can not add right braces '}' as content placeholder.
They will break multiple things, like tabstops for example. Escaping them (`\\}`) does not solve the problem either.

soren-n:
Aim to resolve #60.

To include } in your placeholder content selection, escape it with two \ :
```
it('${1:does something}' ${2:, function($3) {
    $4
\\}});
```

### DIFF
--- a/lib/snippet-body.pegjs
+++ b/lib/snippet-body.pegjs
@@ -2,7 +2,7 @@ bodyContent = content:(tabStop / bodyContentText)* { return content; }
 bodyContentText = text:bodyContentChar+ { return text.join(''); }
 bodyContentChar = escaped / !tabStop char:. { return char; }
 
-escaped = '\\' char:[$\\] { return char; }
+escaped = '\\' char:. { return char; }
 tabStop = simpleTabStop / tabStopWithoutPlaceholder / tabStopWithPlaceholder
 simpleTabStop = '$' index:[0-9]+ {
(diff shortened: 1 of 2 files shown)


lib/snippet-body.pegjs
```diff
@@ -1,25 +1,74 @@
...
-placeholderContentText = text:placeholderContentChar+ { return text.join(''); }
-placeholderContentChar = !tabStop !variable char:[^}] { return char; }
-
-variable = '${' variableContent '}' {
-  return ''; // we eat variables and do nothing with them for now
-}
-variableContent = content:(variable / variableContentText)* { return content; }
-variableContentText = text:variableContentChar+ { return text.join(''); }
-variableContentChar = !variable char:[^}] { return char; }
+start
```
kevinsawicki: Would it be possible to have a more minimal diff here?

The current diff makes it looks like the entire peg was rewritten, yet things like `simpleTabStop` appear to be the same, just reformatted.

soren-n:
Rolled back and now only do the minimal changes

kevinsawicki:
> Rolled back and now only do the minimal changes

Thanks so much for doing this, makes it super easy to review. Also thanks for this fixing this issue :ship:

PR #146, opened 2015-05-29
Load snippets from core packages first

extrabacon:
Snippets from packages are currently loaded according to the order returned from `atom.packages.getLoadedPackages()`. However, core packages are always returned at the end of the list, making it impossible for another package to override snippets from a core package like `language-javascript`.

This pr sorts the results from `getLoadedPackages` so that snippets from core packages are loaded before.

### DIFF
--- a/lib/snippets.coffee
+++ b/lib/snippets.coffee
@@ -123,7 +123,8 @@ module.exports =
 
   loadPackageSnippets: (callback) ->
     packages = atom.packages.getLoadedPackages()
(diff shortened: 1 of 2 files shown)


mnquintana:
Would you mind adding specs for this?

extrabacon:
The build failed because of a timeout on an unrelated spec. What did I do wrong? Everything looks fine on my machine.

benogle:
Was a bunk build. I restarted it, and everything is fine.

This seems reasonable. You can always make the snippets' selectors more specific as well.

PR #147, opened 2015-05-31
:green_heart: Fix load core snippets first spec

lee-dohm:
The spec shouldn't be dependent on the directory from which the snippets are loaded, since that is different depending on whether one is building Atom from source or installing it. Simply check that the first item is from the `language-javascript` package.

This wasn't failing on the snippets package, but is failing for OS X on Core and when run locally.


benogle:
Thanks :heart:

PR #166, opened 2015-09-02
no longer rely on consolidateSelections to remove cursors in spec

lpommers:
Related to the PR on atom core atom/atom#8604

I have a PR on atom core that changes the behavior of consolidateSelections. That build failed because snippets is using that function to remove cursors in its spec.

This PR just removes the cursor manually in the spec - so as not to rely on consolidateSelections


lpommers:
would love to see this merged so I can move on with atom/atom#8604

izuzak:
cc @as-cii since you were looking at atom/atom#8604

as-cii:
Thanks! :zap:

PR #179, opened 2015-10-29
Explicitly mention the period prefix

ethanclevenger91:
Since all the examples use `source.js`, it's unclear if the JS source type has a period in it or if the period is required for all source keys in your snippets file. This makes it very clear, but maybe I'm the only one that was baffled.


benogle:
Thanks

PR #192, opened 2016-02-11
Expansion of nested snippets. Fixes #152.

### LINKED ISSUE #152: Snippets do not behave as expected when expanding within another snippet
Hi,

There is a really nice feature in some editors with snippets where you can expand a snippet at a tab stop inside another snippet, and continue walking through the outer snippet's tab stops after all tab stops in the inner snippet have been visited.  I've tried this in Atom, but it doesn't seem to work, and I can't find any issues describing this as a bug or intentional behavior.  Here is an image of the expected behavior in the Gummi editor:

test

Although the quality of the gif is not that great, you can see that several inner snippets are being expanded and completed at various points while filling in the tab stops of the outer snippet.  In Atom, activating a snippet within another snippet deletes all of the remaining tab stops in the outer snippet once the inner snippet is expanded.  It would be great to have the option of chaining snippets together in this way.

314eter:
snippets

### DIFF
--- a/spec/snippets-spec.coffee
+++ b/spec/snippets-spec.coffee
@@ -651,6 +651,43 @@ describe "Snippets extension", ->
           expect(editor.lineTextForBufferRow(7)).toBe "    }one t1 three"
           expect(editor.lineTextForBufferRow(12)).toBe "};one t1 three"
 
+    describe "when there are nested snippets", ->
+      it "tries to expand before moving to the next tabstop", ->
+        spyOn(Snippets, 'snippetToExpandUnderCursor').andReturn(false)
+        spyOn(Snippets, 'goToNextTabStop').andReturn(true)
+
+        simulateTabKeyEvent()
+        expect(Snippets.snippetToExpandUnderCursor).toHaveBeenCalled()
+        expect(Snippets.goToNextTabStop).toHaveBeenCalled()
+
+      it "moves to surrounding expansions after terminating a snippet", ->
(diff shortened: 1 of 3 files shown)


spec/snippets-spec.coffee
```diff
@@ -655,6 +651,43 @@ describe "Snippets extension", ->
           expect(editor.lineTextForBufferRow(7)).toBe "    }one t1 three"
           expect(editor.lineTextForBufferRow(12)).toBe "};one t1 three"
 
+    describe "when there are nested snippets", ->
+      it "tries to expand before moving to the next tabstop", ->
```
314eter: This test fails. The keybindings are resolved in the wrong order, but only in tests, it works in real life. This is the reason that the next test fails too.

314eter: I'm confused about the current intended behaviour when a snippet is triggered within an existing expansion. According to the specs, it should go to the next tabstop without expanding, but according to the keymaps (snippets-1.cson and snippets-2.cson), it should try to expand first. When I try it myself, it expands most of the times, but not always.

PierBover:
At last!

When will this branch be merged?

stanfieldr:
Any plans to move forward with this?

spec/snippets-spec.coffee
stanfieldr: I think it would make sense to attempt to expand first if possible.

PR #216, opened 2016-10-01
Add an option to disable all included snippets.

CentricStorm:
Add an option to disable all package snippets in one go. This is designed for users that want to start with a clean slate without having to override all of the snippets that they don't want.

### DIFF
--- a/lib/snippets.coffee
+++ b/lib/snippets.coffee
@@ -78,11 +78,14 @@ module.exports =
           @doneLoading()
 
   loadBundledSnippets: (callback) ->
-    bundledSnippetsPath = CSON.resolve(path.join(__dirname, 'snippets'))
-    @loadSnippetsFile bundledSnippetsPath, (snippets) ->
-      snippetsByPath = {}
-      snippetsByPath[bundledSnippetsPath] = snippets
-      callback(snippetsByPath)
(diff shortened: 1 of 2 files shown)


50Wliu:
This setting at the very least needs to be per-package, not an "all-or-nothing" toggle.  You may want to see how the keymap toggle works.

CentricStorm:
> This setting at the very least needs to be per-package, not an "all-or-nothing" toggle.

A separate per-package setting can be added alongside this one.

For some users a per-package setting is suboptimal because:

1. Every time a new package is added, snippets have to be disabled.
2. Every time snippets are added to an existing package, snippets have to be disabled.

For users that want to completely disable all included snippets, both problems simply waste time and increase editor maintenance.

50Wliu:
Your second point is invalid as if the per-package setting is properly implemented, new snippets provided by the package should be automatically disabled as well.

CentricStorm:
> Your second point is invalid as if the per-package setting is properly implemented, new snippets provided by the package should be automatically disabled as well.

If a package starts without _any_ snippets, but then some snippets are added in an update, _those_ snippets _will_ have to be disabled (unless packages are disabled by default, but that doesn't seem likely).

PR #233, opened 2017-02-17
Replace atom-space-pen-views with atom-select-list

as-cii:
This pull request changes `SnippetsAvailable` to use atom-select-list instead of atom-space-pen-views. Feature-wise nothing should have changed, but this is another step toward removing jQuery from Atom.

/cc: @ungb


ungb:
LGTM!  🚢

Issue #55, opened 2014-06-07
Disabling snippets included from packages


kyeweedon:
For example, the `language-javascript` package includes 24 default snippets, many of which I have developed alternatives for which better match my coding style.

How can I block this package from polluting the snippets selection I've defined elsewhere?
# help wanted

jonathandelgado:
+1 - This would really be a fantastic feature.

Fank:
:+1:

pdostal:
:+1:

wyand:
:+1:

Perhaps this could be implemented by an addition to the `snippets.cson` format?

```coffee
'.source.js':
  'setInterval function':
    disabled: true
```

Possible issues on that might be that its inconvenient or unstable to disable by name instead of prefix, and more so the order of operations. Using `disabled` in a package's snippets, instead of in `~/.atom/snippets.cson` or `~/.atom/snippets` could be difficult to manage consistently.

atom-project:
:thumbsup:

varemenos:
:+1: this would really help declutter snippet lists with useless stuff that some of us don't want.

evanwarner:
+1

Janghou:
More Unix-wise:
Why not rename the snippets directory to `snippets_disabled`  or file to `cson_disabled`. It will skip indexing. That's how I do it now manually. 

Just add a setting to rename and tweak the update process to take it in account.

PRO: you have to transparent ways of handling it, manually or through the GUI. 

That's hacking of the good old days.

jprichardson:
Would love this. Just switched over to Atom from Sublime and some of the Atom snippets are super annoying that I'd like to disable but keep the grammar enabled for the sake of syntax highlighting.

benatkin:
:+1:

xicombd:
+1

zyklus:
+1

michaeljacobdavis:
:+1:

oguzbilgic:
+1

oraculus:
:+1:

50Wliu:
This could probably use the same logic as the recent Keybindings PR.

mangecoeur:
+1 - the snippets can be super annoying, for instance if you are editing plain text. The Autocomplete plugin allows you to disable lower priority plugins using a higher priority one which means you can at least muck about with priorities to get the behaviour you want, but the snippets have no such option which can get super annoying.

rearadmiral:
:+1:

[...]

PR #196, opened 2016-03-10
Add a settings view where one can disable snippets for certain packages

dralletje:
I know it does look kind of hacky, but this is the way it should be done I think. I am yet to dive deeper so we can reload without restarting it, maybe be just fully reloading all snippets. Made this because of this issue #55

PR #242, opened 2017-09-15
Add a settings view where one can disable snippets for certain packages + snippets reloading

layonferreira:
This is a continuation of #196 
I have added snippets reloading without the need to restart the editor.
It fixes issue #55

### DIFF
--- a/lib/snippets.coffee
+++ b/lib/snippets.coffee
@@ -129,9 +129,8 @@ module.exports =
 
   loadPackageSnippets: (callback) ->
     # Sort it *before* mapping it
-    packages = atom.packages.getLoadedPackages()
-      .sort (pack, b) ->
-        if /\/app\.asar\/node_modules\//.test(pack.path) then -1 else 1
+    packages = atom.packages.getLoadedPackages().sort (pack, b) ->
+      if /\/app\.asar\/node_modules\//.test(pack.path) then -1 else 1
 
     # So the derived array has the same order
     snippetsDirPaths = (path.join(pack.path, 'snippets') for pack in packages)
@@ -145,38 +144,34 @@ module.exports =
       packagesConfig = {}
       # Add a config entry for every package with snippets
       for o in packagesWithSnippets
-        p = o.pack
-        packagesConfig[p.name] =
-          title: "#{p.name}"
-          description: "Enable snippets for #{p.name}"
+        {pack} = o
+        packagesConfig[pack.name] =
+          title: pack.name
+          description: "Enable snippets for #{pack.name}"
           type: "boolean"
           default: true
 
-      # I don't know where I should get this name from :s
-      # (@name didn't work... even though I had fat arrow)
-      name = 'snippets'
-
       # Update the config schema so you see it in the settings tab
-      atom.config.setSchema(name,
+      atom.config.setSchema('snippets',
         type: 'object',
         properties: packagesConfig
       )
 
       # when snippets package settings changes, reload all package snippets
-      atom.config.onDidChange(name, =>
+      atom.config.onDidChange('snippets', =>
         @unloadPackageSnippets()
         @loadAll()
         @getEmitter().emit 'did-reload-snippets'
       )
 
-      packagesNotDisabled = packagesWithSnippets
+      enabledPackages = packagesWithSnippets
         # Filter out packages disabled by the user
-        .filter (o) -> atom.config.get("#{name}.#{o.pack.name}")
+        .filter (o) -> atom.config.get("snippets.#{o.pack.name}")
         # And map it back to just the results (unzip)
         .map (o) -> o.result
 
       # Pass the enabled packages with snippets forward
-      callback(_.extend({}, packagesNotDisabled...))
+      callback(_.extend({}, enabledPackages...))
 
   doneLoading: ->
     @loaded = true
--- a/lib/snippets.coffee
+++ b/lib/snippets.coffee
@@ -162,19 +162,11 @@ module.exports =
         properties: packagesConfig
       )
 
-      ###
-      TODO:
-      Make it not require a reload to apply config.
-      Right now, when you change the config, it requires you to reload
-      atom as there is no way to remove snippets that belong to a certain package.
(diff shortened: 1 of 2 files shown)


lib/snippets.coffee
```diff
@@ -128,11 +128,55 @@ module.exports =
...
-    snippetsDirPaths = (path.join(pack.path, 'snippets') for pack in packages).sort (a, b) ->
-      if /\/app\.asar\/node_modules\//.test(a) then -1 else 1
-    async.map snippetsDirPaths, @loadSnippetsDirectory.bind(this), (error, results) ->
-      callback(_.extend({}, results...))
+      .sort (pack, b) ->
+        if /\/app\.asar\/node_modules\//.test(pack.path) then -1 else 1
+
+    # So the derived array has the same order
+    snippetsDirPaths = (path.join(pack.path, 'snippets') for pack in packages)
+    async.map snippetsDirPaths, @loadSnippetsDirectory.bind(this), (error, results) =>
```
50Wliu: 💭 I wonder why we're using `async.map` here?  Is the native `Array.map` not performant enough to do synchronously?

Just some thoughts, no need to address them in this PR.

lib/snippets.coffee
```diff
@@ -128,11 +128,55 @@ module.exports =
...
+      # Zip it so we can iterate over it and get the package and result
+      zipped = ({result: result, pack: packages[key]} for key, result of results)
+
+      # Only keep the ones that actually have snippets in them
+      packagesWithSnippets = zipped.filter (o) -> Object.keys(o.result).length isnt 0
+
+      packagesConfig = {}
+      # Add a config entry for every package with snippets
+      for o in packagesWithSnippets
+        p = o.pack
```
50Wliu: How about `{pack} = o`?

lib/snippets.coffee
```diff
@@ -128,11 +128,55 @@ module.exports =
...
+        packagesConfig[p.name] =
+          title: "#{p.name}"
```
50Wliu: This doesn't need interpolation :wink:

lib/snippets.coffee
```diff
@@ -128,11 +128,55 @@ module.exports =
...
+          description: "Enable snippets for #{p.name}"
+          type: "boolean"
+          default: true
+
+      # I don't know where I should get this name from :s
```
50Wliu: Can you elaborate on this comment?

lib/snippets.coffee
```diff
@@ -128,11 +128,55 @@ module.exports =
...
+      )
+
+      # when snippets package settings changes, reload all package snippets
+      atom.config.onDidChange(name, =>
+        @unloadPackageSnippets()
+        @loadAll()
+        @getEmitter().emit 'did-reload-snippets'
+      )
+
+      packagesNotDisabled = packagesWithSnippets
```
50Wliu: A better name might be `enabledPackages`

spec/snippet-loading-spec.coffee
```diff
@@ -1,7 +1,6 @@
 path = require 'path'
 fs = require 'fs-plus'
 temp = require('temp').track()
-
```
50Wliu: Keep this line

spec/snippet-loading-spec.coffee
```diff
@@ -50,6 +49,27 @@ describe "Snippet Loading", ->
...
+    waitsFor "snippets to be changed", ->
+      atom.config.set "snippets.package-with-snippets", true
+      snippet = snippetsService.snippetsForScopes(['.source.js'])['log']
+      snippet?.body is "from-a-community-package"
+
+    runs ->
+      snippet = snippetsService.snippetsForScopes(['.source.js'])['log']
+      expect(snippet.body).toBe "from-a-community-package"
+
+
```
50Wliu: :fire: extra line

iolsen:
Looks like there's solid support for this change. If you can address @50Wliu's feedback we'll get this merged.

layonferreira:
Sure, I will fix the issues after work.

danielbayley:
Should snippets be disabled by default?

layonferreira:
@danielbayley I'm not too sure about that since it would change current behaviour that people are used to.
I believe that this should be introduced as an option to disable and not as an option to enable.
Maybe in the future, if most people are disabling them it would make sense to have it disabled by default.

layonferreira:
Well, it seems that my pr made the specs flicky I will close this PR until I fix it

iolsen:
We're 👍 on the funtionality, @Ben3eeE is going to review and help get this merged.

lib/snippets.coffee
```diff
@@ -28,6 +28,14 @@ module.exports =
 
     snippets = this
 
+    # when snippets package settings changes, reload all package snippets
+    atom.config.onDidChange('snippets', =>
+      if @loaded
+        @unloadPackageSnippets()
+        @loadAll()
```
Ben3eeE: This part worries me:

* We unload package snippets and then `loadAll`. `loadAll` also loads bundled snippets (The snippets included by the snippets pacakge) and user snippets both of which are never unloaded before loading them again.
* I don't think we need to unload all package snippets because one package got disabled or enabled. This needs to change to only affect that package that changed. I'm seeing a significant slowdown when toggling these settings on a slower computer.

lib/snippets.coffee
```diff
@@ -203,6 +246,16 @@ module.exports =
...
+  unloadPackageSnippets: (callback) ->
+    packages = atom.packages.getLoadedPackages()
+    snippetsDirPaths = (path.join(pack.path, 'snippets') for pack in packages).sort (a, b) ->
+      if /\/app\.asar\/node_modules\//.test(a) then -1 else 1
+    async.map snippetsDirPaths, @loadSnippetsDirectory.bind(this), (error, results) =>
+      for snippetSet in results
+        for filepath, _snippetsBySelector of snippetSet
+          @clearSnippetsForPath(filepath)
+
+
```
Ben3eeE: 🔥 Empty line

layonferreira:
Thank you for your review, 
I will look into reimplementing this over the next week.

PR #244, opened 2017-09-30
Make the bundled snippets work when the package is snapshotted

### LINKED ISSUE #243: Builtin snippets are not working when the snippets package is loaded from a V8 snapshot
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

When the snippets package is loaded from a V8 snapshot, I think `__dirname` doesn't point to the correct directory and this code: [lib/snippets.coffee:80-85](https://github.com/atom/snippets/blob/e0909e23e2278419c5465d796ef229d99db21558/lib/snippets.coffee#L80-L85), fails to find the correct path to the bundled `snippets.{cson,json}`.

Might have to pull off something like this [atom/github/lib/helpers.js:7-21](https://github.com/atom/github/blob/98f5ebb2494305549176faa790627a2f3d9f5c4d/lib/helpers.js#L7-L21) for this to work. Though someone might want to refactor so as to reuse the logic rather than copy it at some point.

This is particularly jarring since it is mentioned right there in the [default snippets.cson](https://github.com/atom/atom/blob/1880e1401d7745a2a644c4f7e72cbc625d06023d/dot-atom/snippets.cson) and in [Atom Flight Manual &ndash; Snippets](http://flight-manual.atom.io/using-atom/sections/snippets/).

Note that it does work in dev mode which skips the V8 snapshot

### Steps to Reproduce

1. Open your `snippets.cson` in your `.atom`.
2. Type `snip<tab>`

**Expected behavior:** The builtin "snip" snippet is expanded.

**Actual behavior:** It's not expanded.

**Reproduces how often:** 100%

### Versions

```
Windows 10.0.15063 x64

Atom    : 1.20.1
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0

apm  1.18.4
npm  3.10.10
node 6.9.5 x64
python 2.7.14
git 2.14.1.windows.1
visual studio 2015
```

segevfiner:
### Description of the Change

Add the function `getPackageRoot` and use it to calculate the path to `snippets.{cson,json}`.

I had to manually apply the changes to an Atom clone to test (I want it packaged in asar, snapshotted), by running `script/bootstrap`, copying the changes, and commenting out the bootstrap line in `script/build`. Maybe there is a better way to build Atom with a different private Git copy of a package, but it seems like `packageDependencies` doesn't support Git urls, and there is no obvious way to execute apm to install a Git package into a local Atom Git clone node_modules directory.

### Alternate Designs

Such issues likely repeat them self throughout the builtin Atom packages. Extracting such logic to a shared location and reusing it might help. Note that the exact logic required slightly different from case to case. Some resources are packaged in the asar archive, some are asar.unpacked, some are specially copied in the Atom build scripts. And each such case requires a different function.

### Benefits

Builtin snippets such as `snip` will work again.

### Possible Drawbacks

Yet another "calculate the right path" function.

### Applicable Issues

Fixes #243

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -1,0 +1,13 @@
+/** @babel */
+
+import path from 'path'
+
+export function getPackageRoot() {
+  const {resourcePath} = atom.getLoadSettings()
+  const currentFileWasRequiredFromSnapshot = !path.isAbsolute(__dirname)
+  if (currentFileWasRequiredFromSnapshot) {
+    return path.join(resourcePath, 'node_modules', 'snippets')
+  } else {
+    return path.resolve(__dirname, '..')
+  }
+}
--- a/lib/snippets.coffee
+++ b/lib/snippets.coffee
@@ -8,6 +8,7 @@ ScopedPropertyStore = require 'scoped-property-store'
 
 Snippet = require './snippet'
 SnippetExpansion = require './snippet-expansion'
+{getPackageRoot} = require './helpers'
 
 module.exports =
   loaded: false
@@ -78,7 +79,7 @@ module.exports =
           @doneLoading()
 
   loadBundledSnippets: (callback) ->
-    bundledSnippetsPath = CSON.resolve(path.join(__dirname, 'snippets'))
+    bundledSnippetsPath = CSON.resolve(path.join(getPackageRoot(), 'lib', 'snippets'))
     @loadSnippetsFile bundledSnippetsPath, (snippets) ->
       snippetsByPath = {}
       snippetsByPath[bundledSnippetsPath] = snippets


ungb:
Was able to verify with this change that you can now see `snip` in snippets.cson and prior to this fix it didn't work.  That was the example used in #243

Although I've seen snippets from javascript language package still works before this change.

/cc @as-cii for review.

segevfiner:
> Was able to verify with this change that you can now see snip in snippets.cson and prior to this fix it didn't work. That was the example used in #243
> 
> Although I've seen snippets from javascript language package still works before this change.
> 
> /cc @as-cii for review.

Only the snippets from [lib/snippets.cson](https://github.com/atom/snippets/blob/master/lib/snippets.cson) should be affected by this change.

as-cii:
Nice work, thanks! ⚡️

PR #245, opened 2017-10-02
Correctly place tab stops when there are multiple indented lines

### LINKED ISSUE #140: Tab stops at the start of an indented line select all leading whitespace
This is a pretty specific bug, hopefully I explain it well! 

If you have a snippet with a tab stop that starts at the beginning of a line, it will also select all space to the left of the tab stop. Therefore, if you're attempting to place the snippet within an indented area, the tab stop will also select all the empty space to the left.

_However, this appears to only happen when the snippet is preceded by a multi-line comment._

Below, you'll see two identical snippets, however the first one has a preceding multi-line comment. When tabbing to the `class_name` tab stop, you'll see that it selects too much space, but this is not the case for the second example (the built-in `proto` snippet).

It should be noted that the multi-line comment must be defined as part of the snippet. Writing a multi-line comment, then invoking the `proto` snippet will not cause this bug to occur.

screen shot 2015-05-20 at 10 40 37 am

screen shot 2015-05-20 at 10 41 45 am

My snippet definition:

``` cson
'.source.js'
  'Prototypal method':
    'prefix': 'method'
    'body': """
      /**
       * @method ${1:method_name}
       * @description ${2:description}
       */
      ${3:class_name}.prototype.${1:method_name} = function (${4:first_argument}) {
        ${5:// body...}
      };
    """
```

Here, `${3:class_name}` follows a multi-line comment, and also is also at the start of a line.

### LINKED ISSUE #213: Multi-line snippet not indenting properly
Snippet indentation is incorrect. I have the following snippet:

``` cson
'.source.js':
  'React method':
    'prefix': 'rm',
    'body': """
    $1 = ($2) => {
      $3
    };
    """
```

I open a .js file and type "rm" then hit tab and fill it in.

**Expected behavior:** [What you expected to happen]

Expected result: (e.g.):

``` js
  testMethod = () => {
    this.doSomething();
  };
```

**Actual behavior:** [What actually happened]

``` js
  testMethod = () => {
     this.doSomething();
   };
```
### Versions

You can get this information from executing `atom --version` and `apm --version` at the command line. Also, please include the OS and what version of the OS you're running.

```
Atom    : 1.10.2
Electron: 0.37.8
Chrome  : 49.0.2623.75
Node    : 5.10.0
apm  1.12.5
npm  3.10.5
node 4.4.5
python 2.7.10
git 2.9.3

Mac OS 10.11.6
```

50Wliu:
### Requirements

### Description of the Change

Before this PR, this is how snippet expansion worked:
1. The entire snippet body would be inserted into the editor
2. Tab stop markers would be added
3. Tabs would be converted to spaces (and vice versa) if necessary
4. If the snippet consisted of multiple lines, all lines would be indented to match the first line's indentation level

This process worked pretty well, except for one case - when there was a tab stop at the beginning of a subsequent line in a multiline snippet.  In this case, the tab stop marker would initially be placed at `[line, 0], [line, 0]`.  However, when indenting that line in step 4, the marker is shifted and becomes `[line, 0], [line, indent]` since it touched the area where the shifting occurred (normally, if the editor changes and the marker does not include the changed region, both the start and end get shifted).  As a result, the leading whitespace would be selected instead of just placing the cursor at the tab stop.

With this PR, snippet expansion now works as follows:
1. The indentation level is retrieved and added to each subsequent line of the snippet body if multiline
2. Tab stop markers are shifted to correct for the added indentation
3. Snippet body with indentation is inserted into the editor
4. Tab stop markers corrected for indentation are added
5. Tabs would be converted to spaces (and vice versa) if necessary

### Alternate Designs

This could perhaps be done earlier in the file that calls SnippetExpansion.

### Benefits

I think this logic is more straightforward, since you don't have to think about how the markers will be internally shifted anymore.  It's all done explicitly.

### Possible Drawbacks

I don't see any.

### Applicable Issues

Fixes #140
Fixes #213

### DIFF
--- a/lib/snippet-expansion.coffee
+++ b/lib/snippet-expansion.coffee
@@ -10,17 +10,12 @@ class SnippetExpansion
     @selections = [@cursor.selection]
 
     startPosition = @cursor.selection.getBufferRange().start
-    body = @snippet.body
-    indent = ""
-    tabStopRanges = []
-    if @snippet.lineCount > 1
-      lines = body.split('\n')
-      indent = @editor.lineTextForBufferRow(startPosition.row).match(/^\s*/)[0]
-      for line, index in lines when index isnt 0 # Do not include initial line
-        # Match initial line's indent
-        lines[index] = indent + line
-      body = lines.join('\n')
+    # Get leading indentation level
+    indent = @editor.lineTextForBufferRow(startPosition.row).match(/^\s*/)[0]
+    # Add proper indentation to the snippet
+    body = @snippet.body.replace(/\n/g, '\n' + indent)
 
+    tabStopRanges = []
     for tabStop in @snippet.tabStops
       ranges = []
       for range in tabStop
--- a/lib/snippet-expansion.coffee
+++ b/lib/snippet-expansion.coffee
@@ -1,4 +1,4 @@
-{CompositeDisposable} = require 'atom'
+{CompositeDisposable, Range} = require 'atom'
 
 module.exports =
 class SnippetExpansion
@@ -17,18 +17,19 @@ class SnippetExpansion
       lines = body.split('\n')
       indent = @editor.lineTextForBufferRow(startPosition.row).match(/^\s*/)[0]
       for line, index in lines when index isnt 0 # Do not include initial line
-        # Match first line's indent
+        # Match initial line's indent
         lines[index] = indent + line
       body = lines.join('\n')
 
     for tabStop in @snippet.tabStops
       ranges = []
       for range in tabStop
-        unless startPosition.row is range.start.row
+        newRange = Range.fromObject(range, true) # Don't overwrite the existing range
+        if newRange.start.row # a non-zero start row means that we're not on the initial line
           # Add on the indent offset so that the tab stops are placed at the correct position
-          range.start.column += indent.length
-          range.end.column += indent.length
-        ranges.push(range)
+          newRange.start.column += indent.length
+          newRange.end.column += indent.length
+        ranges.push(newRange)
       tabStopRanges.push(ranges)
 
     @editor.transact =>
--- a/lib/snippet-expansion.coffee
+++ b/lib/snippet-expansion.coffee
@@ -10,11 +10,12 @@ class SnippetExpansion
     @selections = [@cursor.selection]
 
     startPosition = @cursor.selection.getBufferRange().start
+    {body, tabStops} = @snippet
     if @snippet.lineCount > 1 and indent = @editor.lineTextForBufferRow(startPosition.row).match(/^\s*/)[0]
       # Add proper leading indentation to the snippet
-      body = @snippet.body.replace(/\n/g, '\n' + indent)
+      body = body.replace(/\n/g, '\n' + indent)
 
-      tabStops = @snippet.tabStops.map (tabStop) ->
+      tabStops = tabStops.map (tabStop) ->
         tabStop.map (range) ->
           newRange = Range.fromObject(range, true) # Don't overwrite the existing range
           if newRange.start.row # a non-zero start row means that we're not on the initial line
@@ -22,8 +23,6 @@ class SnippetExpansion
             newRange.start.column += indent.length
             newRange.end.column += indent.length
           newRange
-    else
-      {body, tabStops} = @snippet
 
     @editor.transact =>
       newRange = @editor.transact =>
--- a/lib/snippet-expansion.coffee
+++ b/lib/snippet-expansion.coffee
@@ -10,17 +10,36 @@ class SnippetExpansion
     @selections = [@cursor.selection]
 
     startPosition = @cursor.selection.getBufferRange().start
+    body = @snippet.body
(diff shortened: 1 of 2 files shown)


Ingramz:
After fooling around a little, found this in JavaScript:
snippet_broken

lib/snippet-expansion.coffee
```diff
@@ -10,17 +10,37 @@ class SnippetExpansion
...
+    body = @snippet.body
+    indent = ""
+    tabStopRanges = []
+    if @snippet.lineCount > 1
+      lines = body.split('\n')
+      indent = @editor.lineTextForBufferRow(startPosition.row).match(/^\s*/)[0]
+      for line, index in lines when index isnt 0 # Do not include initial line
+        # Match initial line's indent
+        lines[index] = indent + line
+      body = lines.join('\n')
```
Ingramz: `body = @snippet.body.replace('\n', '\n' + indent)`

50Wliu: Can you explain this change?  I'm not sure I follow.

Ingramz: You could replace the code above with
```coffee
indent = @editor.lineTextForBufferRow(startPosition.row).match(/^\s*/)[0]
body = @snippet.body.replace('\n', '\n' + indent)
tabStopRanges = []
```

lib/snippet-expansion.coffee
```diff
@@ -10,17 +10,32 @@ class SnippetExpansion
     @selections = [@cursor.selection]
 
     startPosition = @cursor.selection.getBufferRange().start
+    # Get leading indentation level
+    indent = @editor.lineTextForBufferRow(startPosition.row).match(/^\s*/)[0]
+    # Add proper indentation to the snippet
+    body = @snippet.body.replace(/\n/g, '\n' + indent)
+
+    tabStopRanges = []
```
Ingramz:
```diff
diff --git a/lib/snippet-expansion.coffee b/lib/snippet-expansion.coffee
index bec664d..8902e51 100644
--- a/lib/snippet-expansion.coffee
+++ b/lib/snippet-expansion.coffee
@@ -15,17 +15,14 @@ class SnippetExpansion
     # Add proper indentation to the snippet
     body = @snippet.body.replace(/\n/g, '\n' + indent)

-    tabStopRanges = []
-    for tabStop in @snippet.tabStops
-      ranges = []
-      for range in tabStop
+    tabStopRanges = @snippet.tabStops.map (tabStop) ->
+      tabStop.map (range) ->
         newRange = Range.fromObject(range, true) # Don't overwrite the existing range
         if newRange.start.row # a non-zero start row means that we're not on the initial line
           # Add on the indent offset so that the tab stops are placed at the correct position
           newRange.start.column += indent.length
           newRange.end.column += indent.length
-        ranges.push(newRange)
-      tabStopRanges.push(ranges)
+        newRange

     @editor.transact =>
       newRange = @editor.transact =>
```

Ingramz:
Slightly faster version, avoids grabbing indents from line using regex. Also a little 🐎  in case indents are empty or there is no need for indenting depending on line count.

```diff
diff --git a/lib/snippet-expansion.coffee b/lib/snippet-expansion.coffee
index bec664d..1218564 100644
--- a/lib/snippet-expansion.coffee
+++ b/lib/snippet-expansion.coffee
@@ -10,22 +10,22 @@ class SnippetExpansion
     @selections = [@cursor.selection]

     startPosition = @cursor.selection.getBufferRange().start
-    # Get leading indentation level
-    indent = @editor.lineTextForBufferRow(startPosition.row).match(/^\s*/)[0]
-    # Add proper indentation to the snippet
-    body = @snippet.body.replace(/\n/g, '\n' + indent)
+    if @snippet.lineCount > 1 and startPosition.column > 0
+      # Get leading indentation level
+      indent = @editor.lineTextForBufferRow(startPosition.row).substring(0, startPosition.column)
+      # Add proper indentation to the snippet
+      body = @snippet.body.replace(/\n/g, '\n' + indent)

-    tabStopRanges = []
-    for tabStop in @snippet.tabStops
-      ranges = []
-      for range in tabStop
-        newRange = Range.fromObject(range, true) # Don't overwrite the existing range
-        if newRange.start.row # a non-zero start row means that we're not on the initial line
-          # Add on the indent offset so that the tab stops are placed at the correct position
-          newRange.start.column += indent.length
-          newRange.end.column += indent.length
-        ranges.push(newRange)
-      tabStopRanges.push(ranges)
+      tabStopRanges = @snippet.tabStops.map (tabStop) ->
+        tabStop.map (range) ->
+          newRange = Range.fromObject(range, true) # Don't overwrite the existing range
+          if newRange.start.row # a non-zero start row means that we're not on the initial line
+            # Add on the indent offset so that the tab stops are placed at the correct position
+            newRange.start.column += indent.length
+            newRange.end.column += indent.length
+          newRange
+    else
+      {body, tabStops: tabStopRanges} = @snippet

     @editor.transact =>
       newRange = @editor.transact =>
```

PR #253, opened 2017-10-31
Fix bug where false is never returned from snippetToExpandUnderCursor

leroix:
fix https://github.com/atom/atom/issues/16040


maxbrunsfeld:
Thanks for tracking this down!

leroix:
@50Wliu no worries. Glad it was an easy fix.

nathansobo:
W00t!

No worries @50Wliu. I'm just glad you're keeping things flowing. Mistakes are inevitable.

Issue #257, opened 2017-12-05
Support TextMate-style snippet transformations


savetheclocktower:
This is a gigantic feature request, but I’d argue it’s worth the effort to implement.

Since 1.x, TextMate has supported [snippet “transformations”](https://manual.macromates.com/en/snippets) (see section 7.7) that allow certain tab stops to run a regex substitution on their contents before being put into the buffer. It's an awesome feature and one of the things I dearly miss, since there's no easy way to replicate the functionality.

### Make HTML tag pair

Given a snippet:

```
<${1:p}>$0</${1/\s.*//}>
```

I can expand this snippet, type `div class='row'`, and end up with

```
<div class='row'></div>
```

because the transformation in the latter tab stop replaces the first space and everything after it with an empty string.

### Make Markdown banner

Given a snippet:

```
${1:banner}
${1/./=/}
```

I can expand this snippet, type `Lorem Ipsum`, and end up with

```
Lorem Ipsum
===========
```

because the transformation in the latter tab stop replaces each character I type with a `=` character.

## Challenges

Based on how snippets are currently implemented, I think this would be quite hard. While a snippet is “active,” the snippets package appears to react only to the <kbd>Tab</kbd> key, and doesn’t exert control over how text is inserted. If a numbered tab stop appears in more than one place (“mirroring” in TextMate snippet parlance), the package supports this by creating multiple selections. It would be surprising for a user to see this…

<img width="220" alt="screen shot 2017-12-05 at 2 15 19 pm" src="https://user-images.githubusercontent.com/3450/33629154-b28f355c-d9c8-11e7-9692-863d839ffdf6.png">

…and then, after some typing, find out that those two selections may not be identical.

So to implement this would probably involve hooking into `TextEditor#onDidChange` rather than treating transformed tab-stop content as a selection.

I’d be willing to take a stab at this, but I’m opening this issue to test the waters first. It’s possible I might be the only person on earth who cares about this feature.

savetheclocktower:
Whoops! Duplicate of #40, which was a couple pages back in the issues list.

Issue #40, opened 2014-04-28
Regex Support


Grimeh:
Feature Request:
Support for regular expressions, to enable stuff like method name camel-casing, particularly useful for getters and setters.

postal2600:
Another feature would be to implement smart HTML snippets as in ST: if you type `div#mama` and press `TAB` it will auto-complete to `<div id="mama"></div>`. I would love this in Atom like my next breath. 

If this is implemented it could replace the current snippet list from the html-language.

coopermaruyama:
+1 for this! This is a basic feature in Sublime/Textmate that I really miss.

afanasy:
+1

andersonbarutti:
+1

sdwrage:
+1

ddelrio1986:
+1

killswitch:
:+1:

josa42:
:+1:

rob2022:
:+1:

evanwarner:
+1

rrd108:
+1

zyklus:
+1

prog666:
+1

prog666:
that's the only thing preventing me from using atom every day

anlek:
wow, soul crushed when I learned this.
+:100:

chvonrohr:
:+1:

Stephn-R:
:+1: pinging for updates

wadehammes:
+1

prog666:
-1, learned vim already

PierBover:
+1

sdwrage:
This is not an option for most that don't use vim mode.

jpmurray:
Is there any workaround for this? I'd love to be able to modify the case of some variables in a snippet.

sdwrage:
Not currently that I know of other than just using vim mode. Hopefully
someone will contribute to atom and add the support.

sjmas:
👍

V0idExp:
:+1:

joshmanders:
Thumbs up's aren't needed, Github now has reactions on posts. If you agree with something please use those. We don't need spam in our inboxes with 👍 anymore.

Grimeh:
So would the sublime-style syntax for this be fine?

Instead of the two tab stop forms we have now:
 - `${1}`/`$1`
 - `${1:default}`

We would have an additional form:
 - `${1/regex/format/options}`

In people's opinions, would that be acceptable? 

[...]

PR #260, opened 2017-12-18
Add support for simple regex transformations in tab stops.

savetheclocktower:
_Note: this is my first deep-dive into adding an Atom feature and I’m happy to get pushback on code formatting, architectural choices, and the like._

### Description of the Change

This implements a slimmed-down version of snippet transformations. TextMate’s version of this feature is [documented here](http://manual.macromates.com/en/snippets#transformations).

For example, this snippet:

```
<${1:p}></${1/[ ]+.*$//}>
```

can be used like this:

snippet-transformation-demo

### Design

I went through a bunch of possible ways to design this. Here’s what I landed on:

* Update the `pegjs` file so that we can parse the new syntax.
* Introduce the concept of an `Insertion`, which belongs to a tab stop and can either mirror or transform its input.
* Upon expanding a snippet or moving to a new tab stop, we check if the active tab stop has at least one transformation. If so, we observe the buffer so that every transactional update that happens while that tab stop is active (except for the transactions we initiate) results in a recomputation of the transform output.
* For undo and redo to work the way we want, we have to take the changes that occur as a result of transformations and fold them into the change that kicked off the transformation in the first place. And we need to spy on the history provider so that we can skip trying to reapply transformations upon undo or redo.

### Worth noting

Mirrored tab stops have always been treated like multiple selections…

snippet-mirror mov

…and that doesn’t change. But transformed tab stops are _not_ treated like selections, because I think a user wouldn’t reasonably expect the same input to produce different output when applied to multiple selections.

Transformed tab stops cannot have placeholders; the first transformation gets applied as soon as the snippet is expanded, so any placeholder value would likely be lost immediately.

### What is and isn't included

As explained [in this comment](https://github.com/atom/snippets/issues/40#issuecomment-351204847), I wanted to go minimal here. The find-and-replace syntax roughly corresponds to the two arguments you'd pass to `String#replace` (the regex is automatically given the global flag). There's one extra feature that I brought over from TextMate format strings: the ability to control case through the `\u`, `\l`, `\U`, `\L`, and `\E` flags:

* `\u` and `\l` will force the next character to uppercase or lowercase, respectively.
* `\U` and `\L` will force the rest of the string to uppercase or lowercase, respectively (or until we encounter a different case directive or `\E`).
* `\E` stops the effects of `\U` and `\L`.

There are lots of other features of TextMate format strings, but a lot of them are obscure. I went with this small extension because it's quite useful and is also supported by Sublime Text's snippet transformation syntax. Other features, like conditional insertions, could be added rather easily in the future just by updating the parser and the `Insertion` class.

### Alternate Designs

I don’t like how a snippet expansion needs to listen to `TextBuffer` changes in order to do transformations, but I didn’t see another way. An asynchronous listener doesn’t act quickly enough. But we do listen at the transaction level (`TextBuffer#onDidChangeText`) rather than the keystroke level (`TextEditor#onDidChange`).

If there is a simpler way to do this, I’m all for it.

### Benefits

Makes snippets far more useful. There are a couple more examples in [the (duplicate) ticket that I opened](https://github.com/atom/snippets/issues/257).

### Possible Drawbacks

Not sure. Increased code complexity, perhaps?

**This PR doesn't change any existing behavior** — barely any of the existing tests even needed to be touched — and the theoretical performance penalty of observing `TextBuffer` transactions should only be encountered when using a snippet with a transformation; for existing use cases there should be no performance regression.

### Applicable Issues

Ticket #40 contains the original request.


lib/tab-stop-list.js
```diff
@@ -0,0 +1,46 @@
+/** @babel */
```
nathansobo: It would be nice to avoid using Babel on new files. Everything except the fancy `import` syntax works fine.

lib/snippets.coffee
```diff
@@ -322,15 +330,54 @@ module.exports =
...
+    return if activeExpansions.length is 0 or activeExpansions[0].isIgnoringBufferChanges
+
+    @ignoringTextChangesForEditor editor, ->
+      editor.transact ->
+        for expansion in activeExpansions
+          expansion.textChanged(event)
+
+    # Create a checkpoint here to consolidate all the changes we just made into
+    # the transaction that prompted them.
+    @makeCheckpoint(editor)
```
nathansobo: If I'm reading this right, you're doing this to group any transformations with a previous transaction because the transaction is already closed by the time you receive the event. But I'm confused as to how the initial checkpoint preceding any changes gets populated.

Incidentally, in 1.25 we're adding a `groupLastChanges` method to the buffer that lets you combine the last two transactions on the undo stack, which might make this easier. But you can't really use it until Atom 1.25 hits stable without breaking this package's tests.

lib/snippets.coffee
```diff
@@ -298,6 +299,13 @@ module.exports =
   expandSnippetsUnderCursors: (editor) ->
     return false unless snippet = @snippetToExpandUnderCursor(editor)
 
+    @getStore(editor).observeHistory({
```
nathansobo: We should really supply a flag in the change event indicating when the change is coming from an `undo` or `redo`. If you'd have any interest in adding that, I'd be happy to shepherd it. Then we could simplify this.

spec/insertion-spec.js
```diff
@@ -0,0 +1,141 @@
+/** @babel */
+
+import Insertion from '../lib/insertion';
+import { Range } from 'atom';
+
+const range = new Range(0, 0);
+
+describe('Insertion', () => {
+
```
nathansobo: Super minor, but noticing some extra newlines in the file such as this one.

lib/snippets.coffee
savetheclocktower: Yeah, that’s certainly a better idea. I’ll investigate.

lib/tab-stop-list.js
savetheclocktower: Cool, was unclear on what the best practice was for new files.

lib/snippets.coffee
savetheclocktower: Working off of recollection here, but I think this is the deal:

The initial snippet expansion is easy because it's invoking a command. Everything that we do in response to that command is automatically atomic because of the `TextEditor#transact` call.

The checkpoint logic is needed only for typing that happens while a snippet is active but not in response to a command invocation. Consider: If I have a snippet like `$1\n${1/./\\u$1}` (i.e., transforming what I type to all-caps on the following line):

1. I type its abbreviation and expand the snippet.
2. I type `a`.
3. Atom inserts `A` on the following line.

We're observing text editor changes, so we act _in between_ steps 2 and 3, but we have to bundle the changes of steps 2 and 3 into one atom — i.e., we have to create a transaction that includes something that already happened. The way I arrived at to do this was to make the checkpoint after the initial snippet expansion and then make a new one each time we did step 3. And right before I make a new checkpoint I bundle everything together that happened since the last checkpoint.

This took a lot of time to reason through. I wrote a lot of tests around this precisely because I suspected I didn't understand it correctly.

It does sound like `groupLastChanges` would make this easier — it would theoretically allow me to toss out checkpoints altogether and duct-tape the changes i'm making with the user action that prompted them.

savetheclocktower: After re-reading your question, I'm pretty sure all I did was restate the part that you already understood. The shorter answer to your question is “I don't know.” I’ll investigate when I get a chance.

lib/editor-store.coffee
```diff
@@ -0,0 +1,63 @@
+SnippetHistoryProvider = require './snippet-history-provider'
```
nathansobo: Would you be cool doing this file in pure JS?

lib/snippets.coffee
nathansobo: > The way I arrived at to do this was to make the checkpoint after the initial snippet expansion and then make a new one each time we did step 3

Actually I think this answered my question perfectly.

savetheclocktower: Well, after my long comment I skimmed the source again and can't 100% guarantee that the code does what I claimed it did in the part you quoted. 🤷‍♂️ I can't find the part where I make the initial checkpoint; the `makeCheckpoint` method seems like it only gets called after we set up the listener, not on initial snippet expansion.

nathansobo:
Published as snippets@1.2.0. Thanks for the great work adding this. You combined a bunch of APIs in a pretty complex piece of code to add a really cool feature. I've gone ahead and given you commit bit on this repository ⚡️.

savetheclocktower:
Thanks! I'll revisit this when 1.25 is out, and the undo/redo thing will be on my radar as well.

nobleach:
Forgive my misunderstanding but, is this actually available in non-beta releases? I see it's published in snippets@1.2.0 (current is 1.3.3).

savetheclocktower:
Yeah, it went out with Atom 1.25.0.

nobleach:
That's great news! Looks fantastic. I'll dig around in the source a bit to see how to use it. Thanks!

Issue #266, opened 2018-03-17
Nested tab stops now expand incorrectly


isiahmeadows:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

[Description of the issue]

### Steps to Reproduce

1. Add a snippet with the body `'() => ${1:\\{$2\\}}'`
1. Open a new file
1. Change to a language where that snippet will be in scope
1. Start to type it and hit tab to expand it

**Expected behavior:** [What you expect to happen]

It to expand like this with the brackets selected:

```js
() => {}
```

**Actual behavior:** [What actually happens]

It expands like this, with the brackets selected:

```js
() => {[object Object]}
```

**Reproduces how often:** [What percentage of the time does it reproduce?]

100% of the time

### Versions

You can get this information from copy and pasting the output of `atom --version` and `apm --version` from the command line. Also, please include the OS and what version of the OS you're running.

Atom v1.25.0 on macOS 64-bit. This bug did not exist in v1.24.1.

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

nemoDreamer:
I have a bunch of snippets where a group references another group:

```
'.source.js':
    'Console Log':
        'prefix': 'cl'
        'body': '$3console.${2:log}($1);$0'
    'Console Log (verbose)':
        'prefix': 'cll'
        'body': "$4console.${3:log}('${2:$1}', $1);$0"
    'Console Log (labeled, verbose)':
        'prefix': 'clll'
        'body': "$5console.${4:log}('$1: ${3:$2}', $2);$0"
```

Typing `cll ⇥` and typing something would result in 2 cursors:

```
console.log("myMysteryHash|", myMysteryHash|);
```

But now, it only does

```
console.log("[object Object]", myMysteryHash|);
```

---

[...]

PR #267, opened 2018-04-09
Fix incorrect parsing of placeholders that point to other tab stops.

savetheclocktower:
### Requirements

### Description of the Change

#266 describes a regression with tab stop behavior when the placeholder contains a reference to another tab stop. This almost certainly happened as a result of #260, but there wasn't any test coverage for this particular usage.

### Alternate Designs

The parser was incorrectly assuming that all the items it would receive for a particular rule would be represented as strings, and that it could thus join them all together. Tabstop references are represented as objects. So I wrote a couple of utility functions to ensure that adjacent strings would be merged together without incorrectly coercing objects.

If there's an easier way to do this with a PEG, I'm all for it. Parser grammars are not my strong suit.

### Applicable Issues

#266

### DIFF
--- a/lib/snippet-body.pegjs
+++ b/lib/snippet-body.pegjs
@@ -1,3 +1,27 @@
+{
+  // Joins all consecutive strings in a collection without clobbering any
+  // non-string members.
+  function coalesce (parts) {
+  	var result = [], ri;
+    var part;
+    for (var i = 0; i < parts.length; i++) {
+      part = parts[i];
+      ri = result.length - 1;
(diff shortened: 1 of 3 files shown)


savetheclocktower:
I can merge this myself, but I'll leave it open for a couple days in case other maintainers have feedback.

rsese:
Thanks @savetheclocktower, I'll ask the team to take a look in case anyone has feedback for you :bow:

joefitzgerald:
I have tested and can confirm that this restores the behavior that was present in `v1.24.1`.

Issue #236, opened 2017-06-14
Contiguous tabstops not working as expected (2nd selects part of 1st)


kswedberg:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

Backspacing on second tabstop deletes part of text in first tabstop when tabstops are contiguous.

### Steps to Reproduce

1. Create a simple "foo" snippet, such as:
```cson
  "foo":
    prefix: "foo"
    body: """
    ${2:bar}${3:baz}
    """
```
2. Open a new file and change the syntax to the one the snippet was placed under.
3. Activate the "foo" snippet
4. Type something of 2 characters or more to replace "bar" ("xyzz" for this example)
5. Tab to "baz"

**Expected behavior:** [What you expect to happen]
Only "baz" should be selected for replacement

**Actual behavior:** [What actually happens]
"yzzbaz" is selected for replacement. Only the first character of the first tabstop replacement ("x") is left unselected.

**Reproduces how often:** [What percentage of the time does it reproduce?]
100%

### Versions
Atom 1.17.0
MacOs Sierra 10.12.5

savetheclocktower:
For reference: using `exclusive: true` when we create the marker solves this problem but causes 11 new failures out of the 81 test cases in the spec.

savetheclocktower:
[This discussion](https://github.com/atom-archive/text-document/issues/15) adds some context. Pinging @nathansobo because I’m curious if this ever came up again. I don’t see any open issues in the `text-buffer` repo about it, and `exclusive` is still an all-or-nothing thing — I can’t make the end exclusive but the beginning inclusive.

savetheclocktower:
Thinking about this some more. When the cursor is right at the border of two tab stops (the end of one and the beginning of another) and a character is typed, which marker should “claim” the new text input?

[...]

PR #281, opened 2018-10-24
Fix behavior when inserting content between two adjacent tab stops

### LINKED ISSUE #15: Snippet placeholder bug
I'm trying to write a snippet for `def` in Scala:

```coffee
'method':
  'prefix': 'def'
  'body': 'def ${1:method}${2:(${3:arg}: ${4:Type})} = {\n\t$0\n}'
```

After typing in the method name and hitting tab, the parentheses and arguments should be highlighted, but Atom highlights the most of the method name I typed too (everything but the first character).

savetheclocktower:
Fixes #15.

#### The problem

Here’s a simple example of a snippet that has never quite worked right.

```
${2:bar}${3:baz}
```

I’ll quote from my comment in #236 to explain how this goes wrong:

> I’ll use parentheses to illustrate the invisible markers. After tab expansion, you’ve got:
> 
> ```
> (bar)(baz)
> ```
> 
> You type `x` and now you’ve got:
> 
> ```
> (x)(baz)
> ```
> 
> The editor knows to put `x` into the first marker because bar was selected, and that removed all ambiguity. But right now the cursor is between `x` and `baz`, so when you type text, it could plausibly get added to either marker. We want this…
> 
> ```
> (xyzz)(baz)
> ```
> 
> …but we’re getting this instead:
> 
> ```
> (x)(yzzbaz)
> ```

#### The fix

All snippet markers are getting created with `exclusive: false`. That means that they’ll expand to claim text that’s typed at their boundaries. But when inserted text is right between two adjacent markers, the one on the right always wins. The only way to force it in the direction you want is to create one marker with `exclusive: true` and one with `exclusive: false`.

In our example, `$2` is the active tab stop, so we want its marker to claim the characters that we type. But when we hit <kbd>Tab</kbd> and make `$3` the active tab stop, we want the opposite behavior. So we’ll have to change the settings on the markers as the active tab stop changes — or, rather, copy them to new markers with the settings we want.

There’s one more edge case to consider, and I wasn’t sure how to conceptualize it until I peeked at VScode’s strategy for avoiding this problem: tab stops can reference other tab stops in their placeholder content. If `$2` has a placeholder that includes a reference to `$1`, then `$2`’s marker needs to be inclusive when _either_ `$1` _or_ `$2` is the active tab stop, because we need `$2`’s marker to grow as we type the value for `$1`. If you find this hard to wrap your brain around, you’re not the only one.

### Alternate Designs

The marker API, though robust, is not so rich as to give me lots of options here. The fact that this implementation is nearly identical to VScode’s is validation of the approach.

There were a few different ways I could’ve used to determine the cross-references between tab stops, their placeholders, and any tab stops those placeholders referenced. I decided to capture those references just after parsing, as we consume the syntax tree. Another option would’ve been to inspect the markers after creation and infer the relationships based on which markers were completely surrounded by other markers, but that felt like an indirect approach.

### Benefits

A number of snippets that _seemed_ like they should just work _haven’t_ worked (or have worked incorrectly) for a long time, and now they will.

### Possible Drawbacks

The fact that we have to destroy and recreate markers as tab stops move is not ideal, but I don’t think it’s a bottleneck. If it turns out to be a problem, we can make the changes necessary in text-buffer to allow the `exclusive` setting to be changed after creation on an existing `DisplayMarker`.

### Applicable Issues

 #236 is a duplicate of #15 but contains a lot of relevant discussion.

### DIFF
--- a/lib/insertion.js
+++ b/lib/insertion.js
@@ -45,9 +45,10 @@ function transformText (str, flags) {
 }
 
 class Insertion {
-  constructor ({ range, substitution }) {
+  constructor ({ range, substitution, references }) {
     this.range = range
     this.substitution = substitution
+    this.references = references
     if (substitution) {
       if (substitution.replace === undefined) {
         substitution.replace = ''
--- a/lib/snippet-expansion.coffee
+++ b/lib/snippet-expansion.coffee
@@ -7,7 +7,22 @@ class SnippetExpansion
 
   constructor: (@snippet, @editor, @cursor, @snippets) ->
     @subscriptions = new CompositeDisposable
-    @tabStopMarkers = []
+
+    @insertionsByIndex = []
+    @markersForInsertions = new Map
+
+    # The index of the active tab stop. We don't use the tab stop's own
+    # numbering here; we renumber them consecutively starting at 0 in the order
+    # in which they should be visited. So `$1` will always be index `0` in the
+    # above list, and `$0` (if present) will always be the last index.
+    @tabStopIndex = null
+
+    # If, say, tab stop 4's placeholder references tab stop 2, then tab stop
+    # 4's insertion goes into this map as a "related" insertion to tab stop 2.
+    # We need to keep track of this because tab stop 4's marker will need to be
+    # replaced while 2 is the active index.
+    @relatedInsertionsByIndex = new Map
+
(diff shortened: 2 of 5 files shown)


savetheclocktower:
Closing in favor of #312.

PR #287, opened 2019-02-05
Decaf /lib

Aerijo:
### Requirements

### Description of the Change

Decaffeinates the CoffeeScript files inside the `/lib` directory.

Also fixes a spec that expected 1 console warning, but mine was giving 5 (it didn't say what they were). As the spec tests the message wording however, it should be safe to change `toBe 1` to `toBeGreaterThan 0`

### Alternate Designs

NA

### Benefits

I can understand them now & work on PRs

### Possible Drawbacks

FIrst time decaffeinating. May have missed some safe manual conversions. I did note a lot of `foo != null`, so I wasn't sure if values would be coming in as `null` or `undefined` and left them mostly alone.

Similar with return values (I hate implicit return...). All the specs passed, but that's as much as I can really guarantee. Presumably the return value of a method would be tested if it's important or can vary.


Arcanemagus:
Looks like the AppVeyor container for the `beta` build had networking issues, I restarted the build for that.

lib/snippets.js
```diff
@@ -0,0 +1,675 @@
...
+          On linux there are currently problems with watch sizes. See
+          [this document][watches] for more info.
+          [watches]:https://github.com/atom/atom/blob/master/docs/build-instructions/linux.md#typeerror-unable-to-watch-path\
+          `
+          atom.notifications.addError(message, {dismissable: true})
+        }
+
+        callback(userSnippetsFileDisposable)
+      } else {
+        callback(new Disposable()) // TODO: Work out why it was passed -> in coffeescript
```
Arcanemagus: The CoffeeScript code:
```coffeescript
foo = new Foo
```
Turns into the JavaScript:
```js
foo = new Foo;
```

While the CoffeeScript
```coffeescript
foo = new Foo ->
```
Turns into the JavaScript
```js
foo = new Foo(function() {});
```

----

So it looks like the correct translation here would be:
```js
callback(new Disposable(function () {}))
```

lib/snippets.js
```diff
@@ -0,0 +1,675 @@
...
+      }
+    }
+    return previousTabStopVisited
+  },
+
+  getStore (editor) {
+    return EditorStore.findOrCreate(editor)
+  },
+
+  // createMarkerLayer (editor) {
```
Arcanemagus: 🔥 I'd agree with your commenting out that this is unused code, just remove it 😉.

Aerijo: ... I actually forgot I did that, I was just seeing if anything broke

lib/snippet.js
```diff
@@ -0,0 +1,61 @@
...
+          extractTabStops(content)
+          const range = new Range(start, [row, column])
+
+          const tabStop = this.tabStopList.findOrCreate({
+            index,
+            snippet: this
+          })
+
+          tabStop.addInsertion({ range, substitution })
+
```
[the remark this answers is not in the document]
Aerijo: Yeah, it just felt very crowded with the else clause right below. Whenever I see
```
          tabStop.addInsertion({ range, substitution })
        } else if (typeof segment === 'string') {
          bodyText.push(segment)
          var segmentLines = segment.split('\n')
          column += segmentLines.shift().length
          let nextLine
```
My first thought is "Why is there a random brace hanging off the side of this block?". 

Removed.

lib/snippets.js
Aerijo: That's what it was, but I thought it would be safe to remove the empty function. The disposable should have a working default construction ([and it appears to](https://github.com/atom/event-kit/blob/470fc3f1c433f5b848f7b3a2ae14474521dac422/lib/disposable.js#L35))

Aerijo:
@Arcanemagus I hope you didn't write out every comment each time... I'd have understood even with no comment :smile:

lib/snippet-expansion.js
```diff
@@ -0,0 +1,246 @@
...
+  applyTransformations (tabStop, initial = false) {
+    const items = [...this.tabStopMarkers[tabStop]]
+    if (items.length === 0) { return }
+
+    const primary = items.shift()
+    const primaryRange = primary.marker.getBufferRange()
+    const inputText = this.editor.getTextInBufferRange(primaryRange)
+
+    this.ignoringBufferChanges(() => {
+      for (let index = 0; index < items.length; index++) {
```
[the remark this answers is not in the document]
Arcanemagus: Looks like this got missed in the changes.

lib/snippets.js
Arcanemagus: Wow, somehow I completely glossed over the conditional and only saw the `this.disposalAction()` call when I was looking at that this morning 🤦‍♂️.

Arcanemagus:
Copy/paste meant it wasn't much more than the Ctrl + F to find them 😆.

lib/snippet-expansion.js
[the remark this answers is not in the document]
Arcanemagus: Nevermind, it had selected only your first 7 commits when I was just looking over it.

lib/snippets.js
```diff
@@ -0,0 +1,666 @@
...
+    this.loadSnippetsFile(bundledSnippetsPath, snippets => {
+      const snippetsByPath = {}
+      snippetsByPath[bundledSnippetsPath] = snippets
+      callback(snippetsByPath)
+    })
+  },
+
+  loadUserSnippets (callback) {
+    const userSnippetsPath = this.getUserSnippetsPath()
+    return fs.stat(userSnippetsPath, (error, stat) => {
```
rafeca: this `return` statement seems to not be needed
```suggestion
    fs.stat(userSnippetsPath, (error, stat) => {
```

PR #297, opened 2019-12-05
Decaf the tests

Aerijo:
### Requirements

* Filling out the template is required. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the maintainers' discretion.
* All new code requires tests to ensure against regressions

### Description of the Change

Decaffeinates the remaining CS files in this repo. The functionality of the tests should be unchanged (though some returns were removed, which caused an incorrectly failing test on my machine to pass).

### Alternate Designs

na

### Benefits

Part of the process for introducing the features in #288

### Possible Drawbacks

na

### Applicable Issues

na

PR #299, opened 2019-12-29
Decaf the tests (2nd attempt)

Aerijo:
### Requirements

### Description of the Change

Decaffeinates the remaining CS files in this repo. The functionality of the tests should be unchanged (though some returns were removed, which caused an incorrectly failing test on my machine to pass).

Replaces #297, which ran the decaf before merging upstream changes, and so lost changes in the tests.

### Alternate Designs

na

### Benefits

Part of the process for introducing the features in #288

### Possible Drawbacks

na

### Applicable Issues

na


spec/snippet-loading-spec.js
```diff
@@ -0,0 +1,311 @@
...
+
+      activateSnippetsPackage();
+      runs(() => {
+        atom.config.set('core.packagesWithSnippetsDisabled', ['package-with-snippets']);
+        const allSnippets = snippetsService.getUnparsedSnippets();
+        const scopedSnippet = allSnippets.find(s => s.selectorString === '.package-with-snippets-unique-scope');
+        expect(scopedSnippet).not.toBe(undefined);
+
+        //TODO: This doesn't look right. originalConfig should be restored
+        originalConfig = atom.config.get('core.packagesWithSnippetsDisabled');
```
Aerijo: My linter picked up an unused variable. Based on the others, it should be `atom.config.set('core.packagesWithSnippetsDisabled', originalConfig);` instead

lkashef: Yeah I do agree, this doesn't look right, the setting has been manipulated and we should reset it to the originalConfig, are you running into any issues applying the solution in the comment?

Aerijo: No issue, I just wanted to make sure I should fix it here or a separate PR

lkashef: Normally I would fix it first and then decaf, but it's a big job and the change is simple and obvious and CI is passing, so I think that's alright for now, thanks for the heads up 👍

PR #307, opened 2020-07-14
add windows core package detection

Aerijo:
### Requirements

### Description of the Change

On Windows the snippet file paths use `\` as the delimiter. This doesn't match the `/node_modules/` regex, so community snippets don't get priority.

The change sets the regex depending on if the OS is Windows or not.

### Alternate Designs

None

### Benefits

Priority test passes

### Possible Drawbacks

I don't know why this issue suddenly appeared. It came up in #306, but the cause does not seem to be related to that PR. The check itself has existed (in similar form) for years. The path comes from the Atom package manager module, so perhaps it used to normalise them? Or the files just happened to be in the right order other times? 

### Applicable Issues

None

### DIFF
--- a/lib/snippets.js
+++ b/lib/snippets.js
@@ -11,6 +11,9 @@ const SnippetExpansion = require('./snippet-expansion')
 const EditorStore = require('./editor-store')
 const {getPackageRoot} = require('./helpers')
 
+// Used to guess if the snippets file comes from a core package (bundled with Atom) or not
+const IS_CORE_PACKAGE = process.platform === "win32" ? /\\node_modules\\/ : /\/node_modules\//
+
 module.exports = {
   activate () {
     this.loaded = false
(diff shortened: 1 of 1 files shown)


Aerijo:
The docs have [isBundledPackage](https://flight-manual.atom.io/api/v1.49.0/PackageManager/#instance-isBundledPackage) as a method on the package manager. It may be better to replace the regex test entirely, and use the builtin method instead. I just don't know if using it has already been considered at some point and found to be lacking.

.travis.yml
```diff
@@ -5,6 +5,9 @@ notifications:
     on_success: never
     on_failure: change
 
+env:
```
savetheclocktower, line 8: I'm not familiar enough with the Atom toolchain to know this, so I'll ask: what's the significance of adding this? Is it necessary to fix this issue?

savetheclocktower (changes requested):
I confess to not having looked at this PR until just now, when I realized it'd solve the issue that my PR has with passing CI.

Absent any specific discussion where `isBundledPackage` was considered in the past and rejected, I've got no problem with using it to solve this problem. But the regex test is an improvement over the status quo, so I'd also be cool with leaving it as-is if you don't want to change it for whatever reason.

I left one unrelated question, but I'll be happy to 👍 this once it's answered.

PR #310, opened 2020-09-18
package: adopt standard style

Fluvf:
### Description of the Change
Installs and converts the package to standard style.
While we're at it do some other minor maintenance around `package.json`, see commit messages for details.

### Alternate Designs
N/A

### Benefits
This will help catch any errors in, and keep the style consistent of, any new code.

### Possible Drawbacks
N/A

### Applicable Issues
N/A


Aerijo:
The main repo uses Prettier, it would be best to be consistent with that. [This](https://github.com/atom/atom/blob/master/.prettierrc) is the Prettier config.

Fluvf:
Thanks! I'll convert this pr to use prettier when i get the chance

Fluvf:
I've now added prettier formatting

